PR #36, opened 2022-11-25
Propagate nullability from Kotlin metadata

volivan239:


jcdb-api/src/main/kotlin/org/utbot/jcdb/api/ext/KMetadataExtension.kt
```diff
@@ -52,6 +64,44 @@ val JcMethod.kmConstructor: KmConstructor?
     get() =
         enclosingClass.kMetadata?.constructors?.firstOrNull { it.signature?.name == name && it.signature?.desc == description }
 
+val JcParameter.kmParameter: KmValueParameter?
+    get() {
+        method.kmFunction?.let {
+            // Shift needed to properly handle extension functions
+            val shift = if (it.receiverParameterType != null) 1 else 0
```
lehvolk, line 71: Could you please merge `main` branch and apply fix from there.

jcdb-api/src/main/kotlin/org/utbot/jcdb/api/ext/KMetadataExtension.kt
```diff
@@ -44,6 +52,10 @@ val JcClassOrInterface.kMetadata: KotlinClassMetadata?
         return KotlinClassMetadata.read(kmHeader)
```
lehvolk, line 52: there is no guarantee that this code will not fail. Let's wrap into try-catch and write a warning in logs about it

jcdb-core/src/main/kotlin/org/utbot/jcdb/impl/types/signature/JvmTypes.kt
```diff
@@ -16,54 +16,103 @@
...
+import org.utbot.jcdb.api.ext.isNullable
 
-sealed class JvmType {
+sealed class JvmType(val isNullable: Boolean) {
 
     abstract val displayName: String
 
+    abstract fun setNullability(nullable: Boolean): JvmType
+
+    open fun relaxWithKmType(kmType: KmType): JvmType =
```
lehvolk, line 29: For sure this is shouldn't be interface method of JvmType

jcdb-core/src/main/kotlin/org/utbot/jcdb/impl/types/signature/JvmTypes.kt
```diff
@@ -16,54 +16,103 @@
...
+import kotlinx.metadata.KmType
 import org.utbot.jcdb.api.PredefinedPrimitives
+import org.utbot.jcdb.api.ext.isNullable
 
-sealed class JvmType {
+sealed class JvmType(val isNullable: Boolean) {
 
     abstract val displayName: String
 
+    abstract fun setNullability(nullable: Boolean): JvmType
```
lehvolk, line 27: This method looks like setter but should not be setter. Lets stay with `immutable` pattern naming and do something like for class level api: `notNullable()` or something similar

jcdb-core/src/main/kotlin/org/utbot/jcdb/impl/types/signature/JvmTypes.kt
```diff
@@ -72,6 +121,11 @@ open class JvmTypeVariable(val symbol: String) : JvmType() {
     override val displayName: String
         get() = symbol
 
+    override fun setNullability(nullable: Boolean): JvmType =
+        JvmTypeVariable(symbol, nullable).also {
+            it.declaration = declaration
+        }
+
     override fun equals(other: Any?): Boolean {
```
lehvolk, line 129: Am I right that both types with `nullable = true` and `nullable=false` will be equal?

lehvolk (changes requested):

jcdb-api/src/main/kotlin/org/utbot/jcdb/api/ext/KMetadataExtension.kt
volivan239: Done

jcdb-api/src/main/kotlin/org/utbot/jcdb/api/ext/KMetadataExtension.kt
volivan239: Done

jcdb-core/src/main/kotlin/org/utbot/jcdb/impl/types/signature/JvmTypes.kt
volivan239: Rewrote this logic using extension functions

jcdb-core/src/main/kotlin/org/utbot/jcdb/impl/types/signature/JvmTypes.kt
volivan239: Deleted `equals` and `hashCode` completely as legacy

PR #50, opened 2023-01-26
Support for type annotations (draft)

volivan239:


lehvolk:
IRTest has a serious performance degradation ~30/50% in time. Let's investigate it before merge.

volivan239:
> @volivan239 IRTest has a serious performance degradation ~30/50% in time. Let's investigate it before merge.

I couldn't reproduce this locally. I launched `IRTest` 5 times both on current branch and on main, and what I got was `[48, 33, 39, 44, 47] (~42 average)` seconds for main and `[43, 37, 47, 48, 52] (~45 average)` seconds for current branch

lehvolk:
There is specific test related to Kotlin code `IRTest#get ir of self`. It has degradation but in terms of processing whole test not so much (degradation from 3 sec to 7 sec). I fixed a bit this problem in `main` branch

PR #90, opened 2023-05-31
Add support for approximations mechanism

CaelmBleidd:
In this request, I added support for the approximations mechanism: 

using a provided annotation, you can specify methods and fields that need to be added or overridden in the target class you entered into the annotation. Later, when you access fields and methods of the target class, you get lists of methods and fields where some of the instances replace with virtual ones taken from the approximation class. 

Note: as a user, once you did it, you won't see the initial source code of substituted methods till you use this mechanism. For users all substituted code looks exactly like a real one with the suitable classes, so you can you it without any additional effort.


jacodb-approximations/src/main/kotlin/org/jacodb/approximation/ApproximationsMappingFeature.kt
```diff
@@ -0,0 +1,128 @@
...
+ *  uses it for modifying source code of target classes with information from
+ *  the corresponding approximations.
+ *
+ *  Note: for correct work of the feature, you must wait
+ *  till the whole processing of classpath is finished,
+ *  otherwise you might get incomplete mapping.
+ *  See [JcDatabase.awaitBackgroundJobs].
+ */
+object ApproximationsMappingFeature : JcFeature<Any?, Any?> {
+    private val originalToApproximation: MutableMap<OriginalClassName, ApproximationClassName> = mutableMapOf()
```
lehvolk, line 44: It's better to use CHM since this is global state which is modified from number of threads.

lehvolk:
Let's have some tests here. Then it could be promoted into develop and then immediately released.

lehvolk (changes requested):

lehvolk (approved):

PR #119, opened 2023-07-05
Type inference for equal integral types

### LINKED ISSUE #105: Support type inference for binary operations
Suppose you have the following example:

```java
public int example(short s, char c) {
    return s + c;
}
```

I expect to see a 3-address code in `JcInstList` with type annotations as illustrated below:
```
%0: Int = (arg$0: Char + arg$1: Short): Int
return (%0: Int)
```

But instead I get:
```
%0: Char = (arg$0: Char + arg$1: Short): Char
return (%0: Char)
```
Maybe, the 3-address code builder should infer types for binary expressions based on argument types. E.g., for cases when one of the arguments is `char`, the result should be at least of `int` type.

sergeypospelov:
Fixes #105.

Also adds a test for this case.


lehvolk (approved):

lehvolk:
@sergeypospelov thanks for contribution

PR #155, opened 2023-08-28
Use external distribution of Juliet test suite

Lipen:
This PR proposes to change the vendored Juliet test suite (pre-compiled JAR manually placed in `resources` folder) with an [external distribution](https://github.com/Lipen/juliet-test-suite) via JitPack.


Lipen:
As far as I can tell, all failing tests are [CWE89](https://github.com/Lipen/juliet-test-suite/tree/master/juliet-cwe89/src/main/java/juliet/testcases/CWE89_SQL_Injection), which are NOT present in `juliet.jar` (which was in the repo before the external Juliet suite was included), specifically, s03 and s04 folders.
@volivan239 maybe you can add some "bans" for them, too? Or write a separate analysis for them.

volivan239:
> As far as I can tell, all failing tests are [CWE89](https://github.com/Lipen/juliet-test-suite/tree/master/juliet-cwe89/src/main/java/juliet/testcases/CWE89_SQL_Injection), which are NOT present in `juliet.jar` (which was in the repo before the external Juliet suite was included), specifically, s03 and s04 folders. @volivan239 maybe you can add some "bans" for them, too? Or write a separate analysis for them.

The failing tests are due to defects in the analyzer for SQL injections, I'll create a separate issue for this. Indeed, I haven't tested analyzer on them previously (both for simplicity and saving CI time), let's just ban them for now.

jacodb-analysis/build.gradle.kts
```diff
@@ -6,20 +6,26 @@ plugins {
     kotlin("plugin.serialization") version "1.7.20"
 }
 
+repositories {
```
lehvolk: Please move this to the root build.gradle.kts

jacodb-analysis/build.gradle.kts
```diff
@@ -6,20 +6,26 @@ plugins {
...
 
     testImplementation(testFixtures(project(":jacodb-core")))
     testImplementation(project(":jacodb-api"))
-    testImplementation(group = "javax.servlet", name = "servlet-api", version = "2.5")
     testImplementation(group = "org.junit.jupiter", name = "junit-jupiter-params", version = "5.9.2")
-    testImplementation(files("src/test/resources/juliet.jar"))
     testImplementation(files("src/test/resources/pointerbench.jar"))
     testImplementation(group = "joda-time", name = "joda-time", version = "2.12.5")
+    testImplementation("com.github.Lipen.juliet-test-suite:support:1.3.6")
+    for (cweNum in listOf(89, 476, 563, 690)) {
```
lehvolk, line 20: It's better to store such settings in gradle.properties

jacodb-analysis/build.gradle.kts
```diff
@@ -6,20 +6,26 @@ plugins {
...
     api(project(":jacodb-api"))
 
     testImplementation(testFixtures(project(":jacodb-core")))
     testImplementation(project(":jacodb-api"))
-    testImplementation(group = "javax.servlet", name = "servlet-api", version = "2.5")
     testImplementation(group = "org.junit.jupiter", name = "junit-jupiter-params", version = "5.9.2")
-    testImplementation(files("src/test/resources/juliet.jar"))
     testImplementation(files("src/test/resources/pointerbench.jar"))
     testImplementation(group = "joda-time", name = "joda-time", version = "2.12.5")
+    testImplementation("com.github.Lipen.juliet-test-suite:support:1.3.6")
```
lehvolk: There is a license problem with this dependency. Repo itself doesn't declare any license but have a reference that it is a copy of NIST lib which is under CC 1.0 into README.

I prefer to reference NIST Juliet itself (i.e remove any reference to find-sec-bugs repo) from your repo. Also it's better to host this repo inside UTBot organization.

lehvolk (changes requested):

volivan239 (approved):

jacodb-analysis/build.gradle.kts
Lipen: Why? Jitpack repo is only needed for this subproject.
Is it just to make it easier to keep track of external repos from one place (root build)?

jacodb-analysis/build.gradle.kts
Lipen: Currently, only these CWEs are supported by an existing analysis (afaik), so we only depend on them, but in the future the list will probably grow and might even be replaced with a whole suite.

Gradle properties are just Strings, aren't they? The simple `listOf` in `dependencies {}` seems to be cleaner than parsing `[89, 476, 563, 690]` string into a list of ints.

jacodb-analysis/build.gradle.kts
Lipen: The ZIP archive from NIST does not contain the LICENSE file, either. Should I simply create one (CC0)?

Regarding the repo: I can re-upload the project under UTBot org, without "forking" (I bet I cannot simply "Transfer repo", since it would probably remain to be "forked"), but what about the git history? The initial commits were made by find-sec-bugs's members.

jacodb-analysis/build.gradle.kts
lehvolk: It's a good practice to declare global settings in global place. Even if this settings is used in one submodule.

jacodb-analysis/build.gradle.kts
Lipen: I've moved Juliet repo to https://github.com/UnitTestBot/juliet-java-test-suite

lehvolk:
@Lipen could you please resolve conflicts

lehvolk (approved):

PR #161, opened 2023-08-31
Ifds unit management refactoring #130

### LINKED ISSUE #130: Improve communication between summary and ifds
Current stopping condition for `IfdsInstance` is timeout, which is not flexible. It would be better if Summary somehow would tell IfdsInstance that it has no more facts and can no longer wait for them. Also, current interface of `Summary` is a scratch one, it should be finalized.

volivan239:


jacodb-analysis/src/main/kotlin/org/jacodb/analysis/engine/IfdsUnitRunnerFactory.kt
```diff
@@ -0,0 +1,173 @@
...
+                            QueueEmptinessChanged(backwardQueueIsEmpty && forwardQueueIsEmpty)
+                        manager.handleEvent(newEvent, this@BidiIfdsUnitRunner)
+                    }
+
+                    is SubscriptionForSummaries -> {}
+                }
+            }
+        }
+
+        private var backwardRunner: IfdsUnitRunner<UnitType>? = null
```
lehvolk: Isn't it a `lazy` val?

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/engine/IfdsUnitRunnerFactory.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        override val job: Job = scope.launch(start = CoroutineStart.LAZY) {
+            backwardRunner = backwardRunnerFactory
+                .newRunner(graph.reversed, backwardManager, unitResolver, unit, startMethods, this)
+
+            forwardRunner = forwardRunnerFactory
+                .newRunner(graph, forwardManager, unitResolver, unit, startMethods, this)
+
+            if (isParallel) {
+                backwardRunner!!.job.start()
```
lehvolk: Apart from lazy part it's better to reduce `!!` with local non-nullable val variable

lehvolk (changes requested):

lehvolk (approved):

PR #162, opened 2023-09-01
Skip duplicate CI build on push event in PR

Lipen:
Add condition to skip duplicate build triggered on `pull_request` event on push to PR (on origin repo).

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -24,6 +24,8 @@ permissions:
 jobs:
   jdk11:
 
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


Lipen:
@volivan239 single test is spuriously failing on jdk8 (probably jdk version is irrelevant):
[`at org.jacodb.analysis.impl.NpeAnalysisTest.recursive classes handled correctly(NpeAnalysisTest.kt:631)`](https://github.com/UnitTestBot/jacodb/runs/16414847978)

Note that it passed on `develop` before, and passed on other JDKs just now, so this only indicates that the analysis is not quite deterministic (?).

volivan239:
> @volivan239 single test is spuriously failing on jdk8 (probably jdk version is irrelevant): [`at org.jacodb.analysis.impl.NpeAnalysisTest.recursive classes handled correctly(NpeAnalysisTest.kt:631)`](https://github.com/UnitTestBot/jacodb/runs/16414847978)
> 
> Note that it passed on `develop` before, and passed on other JDKs just now, so this only indicates that the analysis is not quite deterministic (?).

Yes, the tests for IFDS seem to be flaky, it should be fixed in #161

lehvolk:
I just resubmited jobs to be sure

PR #163, opened 2023-09-01
Use matrix on CI

Lipen:
Update `build-and-test` workflow to use matrix strategy, gradle wrapper, and conditions for uploading reports only in one matrix target.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -23,6 +23,7 @@ jobs:
 
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         jdk: [8, 11, 19]
     env:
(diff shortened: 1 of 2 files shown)


Lipen:
Can confirm build failure on local machine with `zulu-8` and `temurin-8` JDKs, investigating.

Lipen:
@lehvolk on JDK 8, the following test finds only 81 usages of `Runnable`:
```kotlin
@Test
fun `find usages of Runnable#run method`() {
    runBlocking {
        val ext = cp.usagesExt()
        val runMethod = cp.findClass<Runnable>().declaredMethods.first()
        assertEquals("run", runMethod.name)
        val result = ext.findUsages(runMethod).toList()
        assertTrue(result.size > 100) { "usages (${result.size}) = $result" }
    }
}
```
(`jacodb-core/src/test/kotlin/org/jacodb/testing/features/BaseSearchUsagesTest.kt:142`)
(last line was modified to show actual usages)

This test does not fail on JDK 11 and 17, so we I assume there are truly more than 100 usages of Runnable in standard library on higher jdks. Probably, previous CI was not failing because it used `jdk+fx` package (for jdk8 and jdk19) which probably contains more than 100 usages of Runnable in total due to inclusion of JavaFX stuff. However, using `jdk+fx` only to satisfy such test is ridiculous. Maybe lower the amount to 50 in this test? If I understand correctly, any number of found usages higher than 0 already proves the correctness of 'findUsages' method.

PR #171, opened 2023-09-05
Publish test results in a separate job

Lipen:
This PR changes CI workflow such that it uploads tests results and jacoco reports as artifacts from all matrix builds, and then publishes them (test results as a comment to PR, and jacoco results to codecov.io) in a separate job

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -3,17 +3,10 @@ name: Build and run tests
 on:
   push:
   pull_request:
-    branches:
-      - develop
+    types: [opened, reopened]
 
 jobs:
   build:
-    # Skip duplicate build when pushing to already existing PR
-    # Note: we decided NOT to skip duplicate builds,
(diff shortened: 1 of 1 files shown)


Lipen:
Test results are being "merged" when combined and published from a separate job. PR comment says "1000 tests" and "3000 runs", which is actually correct, but quite useless. When there are failing tests on some JDK (as it commonly occurs), we are probably not gonna (easily) see which JDK has failed, only that there are "some failing tests among 3000 runs across multiple matrix builds". As far as I understand, one solution is to upload each test result artifact separately, in different job steps, but this will be the same as publishing them immediately after build (and even better - this would not require to upload/download corresponding artifacts with test results/reports).

I'm going to postpone this PR (make it draft) for now, and probably rename and reopen it in the nearest future, since I still want to change workflow event triggers to just `pull_request: types: [opened, reopened]` to eliminate most of duplicate builds.

lehvolk:
Please, check recent changes in `develop`

lehvolk:
Seems outdated. @Lipen feel free to reopen if you have demand for this.

PR #188, opened 2023-10-04
Fix lambda expressions

Saloed:
`InstListBuilder` currenlty uses too wide pattern to recognize lambda and may incorreclty rewrite dynamic invoke.
Also, `JcLambdaExpr` is currently missing important information about the method name and signature.

### DIFF
--- a/jacodb-core/src/main/kotlin/org/jacodb/impl/cfg/JcInstListBuilder.kt
+++ b/jacodb-core/src/main/kotlin/org/jacodb/impl/cfg/JcInstListBuilder.kt
@@ -19,7 +19,9 @@ package org.jacodb.impl.cfg
 import org.jacodb.api.*
 import org.jacodb.api.cfg.*
 import org.jacodb.api.ext.*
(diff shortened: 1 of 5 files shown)


jacodb-core/src/main/kotlin/org/jacodb/impl/cfg/JcInstListBuilder.kt
```diff
@@ -259,29 +261,53 @@ class JcInstListBuilder(val method: JcMethod,val instList: JcInstList<JcRawInst>
     override fun visitJcRawInstanceOfExpr(expr: JcRawInstanceOfExpr): JcExpr =
         JcInstanceOfExpr(classpath.boolean, expr.operand.accept(this) as JcValue, expr.targetType.asType())
 
+    private val lambdaMetaFactory: TypeName by lazy { LAMBDA_METAFACTORY_CLASS.typeName() }
```
lehvolk, line 264: No profit from lazy. Could be made static (i.e moved to companion)

lehvolk (approved):

PR #207, opened 2023-12-08
Added cache for `jcClass`

Damtev:


Saloed (approved):

lehvolk:
@Damtev are there any proves that this change increases performance? 

- Types has less time to live period
- request to find class by name is almost always hits class cache
- JcClass is huge object and will be nice to decouple types and classes. In this PR decopling process will be broken because holding reference to JcClassType almost always will hold link to JcClassOrInterface

Damtev:
@lehvolk `jcClass` is used in calculating the `hashCode` of the `JcClassType`, and we had met a situation when the `hashCode` of an instance of the `JcClassType` has changed. We assumed that happens because the `jcClass` getter sometimes returns different instances of `JcClassOrInterface`, which is quite unexpected for users.

lehvolk:
> @lehvolk `jcClass` is used in calculating the `hashCode` of the `JcClassType`, and we had met a situation when the `hashCode` of an instance of the `JcClassType` has changed.

This is quite strange because JcClassOrInterface designed to have proper equals method. Only case which I can imagine is that the same class located into two different jars. If you check if name and generics substitution is that same then only possible difference may occur into AbstractByteCodeLocation#jarOrFolder file

lehvolk:
@Damtev my concern is that this fix will hide a problem or will not fix the problem.

PR #208, opened 2023-12-11
jacodb-taint-configuration: better `IsType` condition specializer

Saloed:
* Handle the case when package and class matchers are exact after resolving alternatives 
* Faster matching for the whole hierarchy

### DIFF
--- a/jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/TaintConfigurationFeature.kt
+++ b/jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/TaintConfigurationFeature.kt
@@ -16,7 +16,6 @@
 
 package org.jacodb.taint.configuration
 
-import kotlinx.coroutines.runBlocking
 import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.modules.SerializersModule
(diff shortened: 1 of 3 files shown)


CaelmBleidd (approved):
LGTM

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/TaintConfigurationFeature.kt
```diff
@@ -295,31 +293,49 @@ class TaintConfigurationFeature private constructor(
...
-            val disjuncts = mutableListOf<Condition>()
+                position.mapTo(disjuncts) { TypeMatches(it, type) }
+            }
 
-            alternatives.forEach { classMatcher ->
-                val allClasses = runBlocking {
-                    cp.hierarchyExt().findSubClasses(cp.objectClass, allHierarchy = true, includeOwn = true)
+            if (unresolvedMatchers.isNotEmpty()) {
+                val allClassNames = cp.registeredLocations.flatMapTo(hashSetOf()) {
+                    val names = it.jcLocation?.classNames ?: return@flatMapTo emptyList()
```
lehvolk, line 319: This api was not designed for often use so it's better at least to cache it somehow if it's possible

lehvolk:
Just in case almost all new code is not covered by tests

PR #210, opened 2024-01-29
Getting method parameter names

Vassiliy-Kudryashov:
Getting parameter name from "parameters" just doesn't work, this data comes from bytecode ad local variables list.

### DIFF
--- a/jacodb-core/src/main/kotlin/org/jacodb/impl/fs/ByteCodeConverter.kt
+++ b/jacodb-core/src/main/kotlin/org/jacodb/impl/fs/ByteCodeConverter.kt
@@ -114,7 +114,7 @@ private fun MethodNode.asMethodInfo(): MethodInfo {
         parametersInfo = List(params.size) { index ->
             ParameterInfo(
                 index = index,
-                name = parameters?.get(index)?.name,
+                name = argumentName(index),
                 access = parameters?.get(index)?.access ?: Opcodes.ACC_PUBLIC,
                 type = params[index],
                 annotations = visibleParameterAnnotations?.get(index)?.asAnnotationInfos(true).orEmpty()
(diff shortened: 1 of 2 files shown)


jacodb-core/src/main/kotlin/org/jacodb/impl/fs/ByteCodeConverter.kt
```diff
@@ -136,7 +147,7 @@ private fun FieldNode.asFieldInfo() = FieldInfo(
 
 val ClassSource.info: ClassInfo
     get() {
-        return newClassNode(ClassReader.SKIP_CODE).asClassInfo(byteCode)
```
lehvolk: This part is dangerous. Initialization speed needs to be checked.

volivan239:
> Getting parameter name from "parameters" just doesn't work, this data comes from bytecode ad local variables list.

More precisely, correspondence between parameters and local variables table is specified in https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-2.html:

> On class method invocation, any parameters are passed in consecutive local variables starting from local variable 0. On instance method invocation, local variable 0 is always used to pass a reference to the object on which the instance method is being invoked (this in the Java programming language). Any parameters are subsequently passed in consecutive local variables starting from local variable 1.

jacodb-core/src/main/kotlin/org/jacodb/impl/fs/ByteCodeConverter.kt
```diff
@@ -124,6 +124,17 @@ private fun MethodNode.asMethodInfo(): MethodInfo {
...
 
+private fun MethodNode.argumentName(argIndex :Int): String? {
+    localVariables?.let {
+        (argIndex + 1 - (access and Opcodes.ACC_STATIC).countOneBits()).run {
+            if (it.size > this) {
+                return ArrayList(it).sortedBy(LocalVariableNode::index)[this].name
+            }
+        }
+    }
+    return null
```
lehvolk: Seems like fallback `parameters?.get(index)?.name` is missing. Not sure that  localVariables are mandatory for runtime.

Vassiliy-Kudryashov: Thanks, fallback is here now

jacodb-core/src/main/kotlin/org/jacodb/impl/fs/ByteCodeConverter.kt
Vassiliy-Kudryashov: This part was returned back as ClassSource.fullAsmNode does the job when we need some extra information (e.g. names of arguments)

PR #222, opened 2024-02-29
Flow-less summary storage

Lipen:
This PR proposes to add a summary storage without Flow's, which allows for deterministic (non-asynchronous!) subscriptions on summary edges.
Ultimately, this enables us to use `MethodUnitResolver` without any doubts, which currently is not very possible to due the fact that method-bound runners might die (i.e. might be killed by the manager) due to their queue exhaustion before the subscription on summary edges will be handled by the separate spawned coroutine (`launch` + `SharedFlow::collect`).

TODO:
- consider renaming the existing summary storage impl
- refine the interface of the summary storage
- add tests utilizing the `MethodUnitResolver`


Lipen:
Moved to https://github.com/UnitTestBot/usvm/pull/235

PR #223, opened 2024-03-05
Add documentation for taint configuration

CaelmBleidd:


jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/Position.kt
```diff
@@ -23,31 +23,55 @@ fun interface PositionResolver<out R> {
...
+ * It is for regularly returned objects only, and it is not suitable for thrown exceptions.
+ */
 @Serializable
 @SerialName("Result")
 object Result : Position {
     override fun toString(): String = javaClass.simpleName
 }
 
+/**
+ * Represents any element of the collection (string, array, list, etc.),
```
Lipen, line 72: This description is "technically correct", but is not very helpful to mirror the reality and current implementation. We only interpret `ResultAnyElement` position as "any element of an array". Currently, it is not possible to express "all elements of a collection (e.g. List) are tainted, but the collection itself is not" using this Position alone — for collections, we simply over-approximate it via `PassThrough` rules, e.g. on `List::get`: `CopyAllMarks(This, Result)`, without relying on `ResultAnyElement`.

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/Position.kt
```diff
@@ -23,31 +23,55 @@ fun interface PositionResolver<out R> {
...
 
+/**
+ * A class representing a position of tainted data in a method.
+ */
 @Serializable
 sealed interface Position
 
+/**
+ * Represents an argument of a method call.
+ * Numeration starts from zero, `this` parameter is not included. <p>
```
Lipen: `<p>`?

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/Position.kt
```diff
@@ -23,31 +23,55 @@ fun interface PositionResolver<out R> {
...
+ */
 @Serializable
 @SerialName("Result")
 object Result : Position {
     override fun toString(): String = javaClass.simpleName
 }
 
+/**
+ * Represents any element of the collection (string, array, list, etc.),
+ * returned as a result from a method.
```
Lipen: "from a method CALL"

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/Position.kt
```diff
@@ -23,31 +23,55 @@ fun interface PositionResolver<out R> {
     fun resolve(position: Position): R
 }
 
+/**
+ * A class representing a position of tainted data in a method.
```
Lipen: This is not a class

Maybe simply "A representation of a position of tainted data"?

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/TaintAction.kt
```diff
@@ -46,6 +46,13 @@ val actionModule = SerializersModule {
 
 // TODO add marks for aliases (if you pass an object and return it from the function)
 
+/**
+ * Copies all the marks from the [from] object to the [to].
+ * As a result, all marks from the initial object will be copied into the target object.
+ * If some marks were present only in the [to] object, they remain unchanged. <p>
```
Lipen: `<p>`?

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/Position.kt
```diff
@@ -23,31 +23,55 @@ fun interface PositionResolver<out R> {
...
+ * Represents `this` argument of a method call.
+ */
 @Serializable
 @SerialName("This")
 object This : Position {
     override fun toString(): String = javaClass.simpleName
 }
 
+/**
+ * Represents a resulting value of a method call.
```
Lipen:
```suggestion
 * Represents the resulting value of a method call.
```

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/Position.kt
```diff
@@ -23,31 +23,55 @@ fun interface PositionResolver<out R> {
...
+ * A class representing a position of tainted data in a method.
+ */
 @Serializable
 sealed interface Position
 
+/**
+ * Represents an argument of a method call.
+ * Numeration starts from zero, `this` parameter is not included. <p>
+ *
+ * For instance, `obj.foo(a, b)` -> a := Argument(0), b := Argument(1)
```
Lipen:
```suggestion
 * For instance, `obj.foo(a, b)` -> `a := Argument(0)`, `b := Argument(1)`
```

Lipen (changes requested):
Looks good, but there are some minor issues: feel free to resolve conversations if you believe they should not be fixed as proposed.

Regarding `<p>` — maybe I'm not aware, what does it do in KDoc?

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/TaintAction.kt
CaelmBleidd: It is a way to create a line break in the docs: both HTML and idea rendered

image

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/Position.kt
CaelmBleidd: The configuration doesn't depend on the current implementation since there are at least two different implementations: here and in USVM. So, the question is "What is this supposed to mean in the perfect world?". As far as I understood, this is exactly what is written -- position somewhere in the array: any element of the array might be the one.

jacodb-taint-configuration/src/main/kotlin/org/jacodb/taint/configuration/TaintAction.kt
Lipen: (resolved in person)
Resume:

`<p>` is not necessary, `\n\n` (two enters) are enough in the modern software.

```kotlin
/**
 * Test line 1.
 *
 * Test line 2.
 */
fun test() {}
```

dokka HTML:
image

Lipen:
@CaelmBleidd for `dokkaHtmlMultiModule` to work, we also need some `README.md` (`dokka` is fine even with an empty `touch`-ed file) in `jacodb-taint-configuration` module. Otherwise, it fails (at least, locally). This is gonna be necessary when we are going to release.

Lipen (approved):
Nice!

PR #225, opened 2024-04-04
IFDS core refactor

sergeypospelov:
# Description

## PR TODOs:
- [x] Unused Analysis
- [x] Multi chunk scenario + tests
- [x] Graceful shutdown on timeout
- [x] Try to remove `Fact` generic parameter
- [x] Revert Kotlin version (better fix `IRTest`)
- [x] Unify helper extension functions (system creation, analysis start, result retrieval)
- [x] Fix workaround with `flag`
- [x] Remove and replace old entities
- [x] Some tests
- [ ] Comments
- [ ] Description


jacodb-ifds/actors/src/main/kotlin/org/jacodb/actors/impl/ActorSystemImpl.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    companion object {
+        private const val USER_ACTOR_NAME = "usr"
+        private const val WATCHER_ACTOR_NAME = "watcher"
+    }
+}
+
+// TODO: consider just 'system', because 'Of' implies a vararg, which is not the case here
+fun <Message> systemOf(
```
Lipen: Consider renaming `systemOf` to just `system`, because 'Of' implies a vararg argument, which is not the case here.

jacodb-analysis/src/main/kotlin/org/jacodb/ifds/ChunkStrategies.kt
```diff
@@ -0,0 +1,61 @@
...
+package org.jacodb.ifds
+
+import org.jacodb.api.JcClassOrInterface
+import org.jacodb.api.JcMethod
+import org.jacodb.api.cfg.JcInst
+import org.jacodb.api.ext.packageName
+import org.jacodb.ifds.domain.Chunk
+
+// TODO: consider 'SingletonChunk'
+data object SingleChunk : Chunk
```
Lipen: Consider renaming `SingleChunk` to `SingletonChunk`. (in fact, SingleChunk is also okay, feel free to resolve -- but remove the TODO).

jacodb-ifds/ifds/src/main/kotlin/org/jacodb/ifds/result/Merging.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package org.jacodb.ifds.result
+
+import org.jacodb.ifds.domain.Edge
+import org.jacodb.ifds.domain.Reason
+import org.jacodb.ifds.domain.Vertex
+
+fun <Stmt, Fact, Result : IfdsResult<Stmt, Fact>> mergeIfdsResults(
+    ifdsDatas: Collection<IfdsComputationData<Stmt, Fact, Result>>,
```
Lipen: "data" is already a plural form of "datum", so "datas" does not make sense :)

settings.gradle.kts
```diff
@@ -27,3 +27,5 @@ include("jacodb-benchmarks")
 include("jacodb-cli")
 include("jacodb-approximations")
 include("jacodb-taint-configuration")
+include("jacodb-ifds:actors")
+include("jacodb-ifds:ifds")
```
Lipen: If you are proposing to start using nested modules (which might be a pain to use in some cases), why not put "actors" and "ifds" as submodules of "jacodb-analysis"? That is, consider making a single module "jacodb-analysis" (it is already here), move everything from it to "jacodb-analysis:taint" (as it is taint-specific), and add your two modules to it. Or better just put everything into one gradle module "jacodb-analysis" - what is the value of splitting the codebase into nested submodules here?

jacodb-ifds/ifds/src/main/kotlin/org/jacodb/ifds/messages/CommonMessages.kt
```diff
@@ -0,0 +1,25 @@
...
+ *  limitations under the License.
+ */
+
+package org.jacodb.ifds.messages
+
+import org.jacodb.ifds.domain.RunnerId
+
+sealed interface CommonMessage
+
+sealed interface RunnerMessage : CommonMessage {
```
Lipen: Consider moving RunnerMessage into a separate file

jacodb-ifds/ifds/src/main/kotlin/org/jacodb/ifds/domain/FlowFunction.kt
```diff
@@ -0,0 +1,37 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.jacodb.ifds.domain
+
+// TODO: consider 'FlowFunctions' (plural)
+interface FlowFunction<Stmt, Fact> {
```
Lipen: Consider renaming it to `FlowFunctions` (plular)

Lipen: Consider adding `obtainStartFacts(method)` "flow function" (here, or inside analyzer(?))

jacodb-analysis/src/test/kotlin/org/jacodb/analysis/impl/JodaDateTimeAnalysisTest.kt
```diff
@@ -53,37 +64,66 @@ class JodaDateTimeAnalysisTest : BaseTest() {
...
+
+        for (method in methods) {
+            system.startNpeAnalysis(method)
+        }
+        launch {
+            delay(20.seconds)
+            system.stop()
+        }
+        system.awaitCompletion()
+        system.resume()
```
Lipen: Why do you need to call `resume` here? Other tests (juliet-sql, for example) do not contain such calls. Is it because of `system.stop()` above? In this case, why not implement the timeout via `launch { 
delay(); stop(); resume() }`, i.e. call `resume` after `stop` only when necessary?

jacodb-analysis/src/main/kotlin/org/jacodb/ifds/taint/Runners.kt
```diff
@@ -0,0 +1,22 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.jacodb.ifds.taint
+
+import org.jacodb.ifds.domain.RunnerId
+
+data object ForwardRunner : RunnerId
```
Lipen: Consider renaming to `ForwardRunnerId`

PR #232, opened 2024-07-04
Switch ArkTS implementation to ARK IR

CaelmBleidd:


jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/EtsTraits.kt
```diff
@@ -0,0 +1,160 @@
...
+import org.jacodb.analysis.util.toPathOrNull as _toPathOrNull
+import org.jacodb.panda.dynamic.ets.utils.getOperands as _getOperands
+import org.jacodb.panda.dynamic.ets.utils.getValues as _getValues
+
+interface EtsTraits : Traits<EtsMethod, EtsStmt> {
+
+    override val CommonCallExpr.callee: EtsMethod
+        get() {
+            check(this is EtsCallExpr)
+            // return cp.getMethodBySignature(method) ?: error("Method not found: $method")
```
CaelmBleidd: Commented code

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/EtsTraits.kt
```diff
@@ -0,0 +1,160 @@
...
+            return EtsMethodImpl(method)
+        }
+
+    override val EtsMethod.thisInstance: EtsThis
+        get() = EtsThis(EtsAnyType)
+
+    override val EtsMethod.isConstructor: Boolean
+        get() = false
```
CaelmBleidd: Why?

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/EtsTraits.kt
```diff
@@ -0,0 +1,160 @@
...
+        return method.parameters.map { getArgument(it) }
+    }
+
+    override fun CommonValue.isConstant(): Boolean {
+        check(this is EtsEntity)
+        return this is EtsConstant
+    }
+
+    override fun CommonValue.eqConstant(constant: ConstantValue): Boolean {
+        TODO("Not yet implemented")
```
CaelmBleidd, line 106: Why? Do we need it?

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/EtsTraits.kt
```diff
@@ -0,0 +1,160 @@
...
+        return _getValues().toSet()
+    }
+
+    override fun EtsStmt.getOperands(): List<EtsEntity> {
+        return _getOperands().toList()
+    }
+
+    companion object : EtsTraits {
+        // Note: unused for now
+        // lateinit var cp: EtsFile
```
CaelmBleidd: When will we use it? Maybe we can remove it for now?

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/PandaStaticTraits.kt
```diff
@@ -178,7 +174,10 @@ interface PandaStaticTraits : Traits<PandaMethod, PandaInst> {
     }
 
     // Ensure that all methods are default-implemented in the interface itself:
-    companion object : PandaStaticTraits
+    companion object : PandaStaticTraits {
+        // Note: unused for now
+        // lateinit var cp: PandaProject
```
CaelmBleidd, line 179: Same

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsConstant.kt
```diff
@@ -0,0 +1,156 @@
...
+    val elements: List<EtsEntity>,
+    override val type: EtsType, // EtsArrayType
+) : EtsConstant {
+
+    // EtsArrayType
+    // init {
+    //     require(type.dimensions == 1) {
+    //         "Array type of array literal must have exactly one dimension"
+    //     }
+    // }
```
CaelmBleidd: Commented code

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsConstant.kt
```diff
@@ -0,0 +1,156 @@
...
+    override fun toString(): String {
+        return elements.joinToString(prefix = "[", postfix = "]")
+    }
+
+    override fun <R> accept(visitor: EtsConstant.Visitor<R>): R {
+        return visitor.visit(this)
+    }
+}
+
+// TODO: replace `Pair<String, Value>` with `Property`
```
CaelmBleidd: Do it?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsExpr.kt
```diff
@@ -0,0 +1,309 @@
...
+    override fun <R> accept(visitor: EtsExpr.Visitor<R>): R {
+        return visitor.visit(this)
+    }
+}
+
+data class EtsNewArrayExpr(
+    val elementType: EtsType,
+    val size: EtsEntity,
+) : EtsExpr {
+    // TODO: support multi-dimensional arrays
```
CaelmBleidd: Throw an error if you get size other than one

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsExpr.kt
```diff
@@ -0,0 +1,309 @@
...
+//     val trueBranch: Value,
+//     val falseBranch: Value,
+// ) : Expr {
+//     override val type: Type
+//         get() = TypeInference.commonType(trueBranch.type, trueBranch.type)
+//
+//     override fun toString(): String {
+//         return "$condition ? $trueBranch : $falseBranch"
+//     }
+// }
```
CaelmBleidd: Commented code

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsExpr.kt
```diff
@@ -0,0 +1,309 @@
...
+// TODO: AddExpr and many others
+// data class AddExpr(
+//     override val left: Value,
+//     override val right: Value,
+// ) : BinaryExpr {
+//     override fun toString(): String {
+//         return "$left + $right"
+//     }
+// }
+
```
CaelmBleidd: Do we need it?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsExpr.kt
```diff
@@ -0,0 +1,309 @@
...
+// data class ConditionExpr(
+//     val relop: String,
+//     override val left: Value,
+//     override val right: Value,
+// ) : BinaryExpr {
+//     override fun toString(): String {
+//         return "$left $relop $right"
+//     }
+// }
```
CaelmBleidd: Same

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/Ops.kt
```diff
@@ -0,0 +1,275 @@
...
+
+    /**
+     * `!=`
+     */
+    NotEq,
+
+    /**
+     * `===`
+     */
+    EqEqEq,
```
CaelmBleidd: StrictEq?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/Ops.kt
```diff
@@ -0,0 +1,275 @@
...
+
+    /**
+     * `!==`
+     */
+    NotEqEq,
```
CaelmBleidd: NotStrictEq

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/Ops.kt
```diff
@@ -0,0 +1,275 @@
...
+
+    /**
+     * `<`
+     */
+    Lt,
+
+    /**
+     * `<=`
+     */
+    LtEq,
```
CaelmBleidd: Le, Ge

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/Ops.kt
```diff
@@ -0,0 +1,275 @@
...
+enum class AssignOp {
+    /**
+     * '='
+     */
+    Assign,
+
+    /**
+     * '+='
+     */
+    AddAssign,
```
CaelmBleidd: plusassign to match with kotlin?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/Position.kt
```diff
@@ -0,0 +1,83 @@
...
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.jacodb.panda.dynamic.ets.base
```
CaelmBleidd: Shouldn't we rename it just to `org.jacodb.ets`?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/Visitors.kt
```diff
@@ -0,0 +1,199 @@
...
+    override fun visit(value: EtsNullConstant) {
+        println("visit<${value::class.java.simpleName}>(value = $value)")
+    }
+
+    override fun visit(value: EtsUndefinedConstant) {
+        println("visit<${value::class.java.simpleName}>(value = $value)")
+    }
+}
+
+fun main() {
```
CaelmBleidd: What is this main for?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Model.kt
```diff
@@ -0,0 +1,140 @@
...
+    val projectDir: String? = null,
+    val projectName: String? = null,
+    val namespaces: List<NamespaceDto>,
+    val classes: List<ClassDto>,
+    val importInfos: List<ImportInfoDto>,
+    val exportInfos: List<ExportInfoDto>,
+) {
+    companion object {
+        private val json = Json {
+            // classDiscriminator = "_"
```
CaelmBleidd: Do we need it?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Model.kt
```diff
@@ -0,0 +1,140 @@
...
+    val isDefault: Boolean,
+    val modifiers: List<String>,
+    val originTsPosition: LineColPositionDto,
+)
+
+// @Serializable
+// data class Decorator(
+//     val kind: String,
+//     val type: String?,
+// )
```
CaelmBleidd: ?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Values.kt
```diff
@@ -0,0 +1,421 @@
...
+}
+
+// @Serializable
+// @SerialName("UNKNOWN")
+// data class UnknownValueDto(
+//     val value: JsonElement,
+// ) : ValueDto {
+//     override val type: TypeDto
+//         get() = UnknownTypeDto
+// }
```
CaelmBleidd: Commented code

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Values.kt
```diff
@@ -0,0 +1,421 @@
...
+) : ImmediateDto {
+    override fun toString(): String {
+        return "'$value': $type"
+    }
+}
+
+// @Serializable
+// @SerialName("StringConstant")
+// data class StringConstant(
+//     val value: String,
```
CaelmBleidd: What it all this commented code for?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Values.kt
```diff
@@ -0,0 +1,421 @@
...
+    }
+}
+
+@Serializable
+@SerialName("NewArrayExpr")
+data class NewArrayExprDto(
+    val elementType: TypeDto,
+    val size: ValueDto,
+) : ExprDto {
+    // TODO: support multi-dimensional arrays, when upstream implement it
```
CaelmBleidd: Throw notimplementederror then

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Values.kt
```diff
@@ -0,0 +1,421 @@
...
+        return elements.joinToString(prefix = "[", postfix = "]")
+    }
+}
+
+@Serializable
+@SerialName("ObjectLiteralExpr")
+// TODO: keys and values
+data class ObjectLiteralDto(
+    // val keys: List<String>,
+    // val values: List<Value>,
```
CaelmBleidd: Why is it commented with TODO?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Values.kt
```diff
@@ -0,0 +1,421 @@
...
+}
+
+@Serializable
+sealed interface BinaryExprDto : ExprDto {
+    val left: ValueDto
+    val right: ValueDto
+
+    // override val type: TypeDto
+    //     get() = UnknownTypeDto
+}
```
CaelmBleidd: Commented code

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Values.kt
```diff
@@ -0,0 +1,421 @@
...
+) : BinaryExprDto {
+    override fun toString(): String {
+        return "$left $op $right"
+    }
+}
+
+@Serializable
+sealed interface ConditionExprDto : BinaryExprDto {
+    // override val type: String
+    //     get() = "boolean"
```
CaelmBleidd: Same

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
```diff
@@ -0,0 +1,712 @@
+/*
```
CaelmBleidd: Please, name files starting with a capital letter

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
```diff
@@ -0,0 +1,712 @@
...
+    }
+
+    fun convertToEtsStmt(stmt: StmtDto) {
+        val newStmt = when (stmt) {
+            is UnknownStmtDto -> object : EtsStmt {
+                override val location: EtsInstLocation = loc()
+
+                override fun toString(): String = "Unknown(${stmt.stmt})"
+
+                // TODO: equals/hashCode ???
```
CaelmBleidd: Implement it?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
```diff
@@ -0,0 +1,712 @@
...
+        }
+        currentStmts += newStmt
+    }
+
+    fun convertToEtsEntity(value: ValueDto): EtsEntity {
+        return when (value) {
+            is UnknownValueDto -> object : EtsEntity {
+                override val type: EtsType = EtsUnknownType
+
+                // override fun toString(): String = "Unknown(${value.value})"
```
CaelmBleidd: Rmove it

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
```diff
@@ -0,0 +1,712 @@
...
+            )
+
+            is CastExprDto -> EtsCastExpr(
+                arg = convertToEtsEntity(value.arg),
+                type = convertToEtsType(value.type),
+            )
+
+            is PhiExprDto -> EtsPhiExpr(
+                args = value.args.map { convertToEtsEntity(it) },
+                argToBlock = emptyMap(), // TODO
```
CaelmBleidd: Why is it not implemented?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
```diff
@@ -0,0 +1,712 @@
...
+                type = convertToEtsType(value.type),
+            )
+
+            is ArrayLiteralDto -> EtsArrayLiteral(
+                elements = value.elements.map { convertToEtsEntity(it) },
+                type = convertToEtsType(value.type), // TODO: as EtsArrayType,
+            )
+
+            is ObjectLiteralDto -> EtsObjectLiteral(
+                properties = emptyList(), // TODO
```
CaelmBleidd: Same

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
```diff
@@ -0,0 +1,712 @@
...
+                right = convertToEtsEntity(value.right),
+            )
+
+            is InstanceCallExprDto -> EtsInstanceCallExpr(
+                instance = convertToEtsEntity(value.instance),
+                method = convertToEtsMethodSignature(value.method),
+                args = value.args.map {
+                    val etsEntity = convertToEtsEntity(it)
+                    if (etsEntity is EtsValue) return@map etsEntity
+                    val newLocal = EtsLocal("_tmp${freeLocal++}", EtsUnknownType)
```
CaelmBleidd: Please, use empty lines to separate different code blocks

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
```diff
@@ -0,0 +1,712 @@
...
+                }
+            }
+        }
+
+        val successorMap: MutableMap<EtsStmt, List<EtsStmt>> = hashMapOf()
+        for (block in cfg.blocks) {
+            if (block.stmts.isEmpty()) {
+                continue
+            }
+            val startId = blockStart[block.id]!!
```
CaelmBleidd: getValue insted of !!

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
```diff
@@ -0,0 +1,712 @@
...
+                                    type = ClassTypeDto(classSignatureDto)
+                                )
+                            )
+                        )
+                    )
+                )
+            )
+        )
+    )
+
```
CaelmBleidd: Please, construct these objects separately and create without soo many nested levels of instructions

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
```diff
@@ -0,0 +1,712 @@
...
+        UnknownTypeDto -> EtsUnknownType
+
+        VoidTypeDto -> EtsVoidType
+    }
+}
+
+// TODO: add EtsNeverConstant?
+// TODO: add EtsUnknownConstant(value: String)?
+fun convertToEtsConstant(value: ConstantDto): EtsConstant {
```
CaelmBleidd: Please, check all todos and commentd code in the requirest

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/graph/EtsCfg.kt
```diff
@@ -0,0 +1,140 @@
...
+        for ((stmt, nexts) in successorMap) {
+            for (next in nexts) {
+                map.getOrPut(next) { hashSetOf() } += stmt
+            }
+        }
+        map
+    }
+
+    override fun throwers(node: EtsStmt): Set<EtsStmt> {
+        TODO("Current version of IR does not contain try catch blocks")
```
CaelmBleidd: Is it still true?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/graph/EtsCfg.kt
```diff
@@ -0,0 +1,140 @@
...
+
+fun EtsCfg.findDominators(): GraphDominators<EtsStmt> {
+    return GraphDominators(this).also { it.find() }
+}
+
+fun EtsCfg.view(dotCmd: String, viewerCmd: String, viewCatchConnections: Boolean = false) {
+    Util.sh(arrayOf(viewerCmd, "file://${toFile(dotCmd, viewCatchConnections)}"))
+}
+
+fun EtsCfg.toFile(dotCmd: String, viewCatchConnections: Boolean = false, file: File? = null): Path {
```
CaelmBleidd: Extract thus function to some separate util file, it is not a part of the logic for the EtsCfg.kt

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/graph/EtsCfg.kt
```diff
@@ -0,0 +1,140 @@
...
+        val node = Node("$index")
+            .setShape(Shape.box)
+            .setLabel(inst.toString().replace("\"", "\\\""))
+            .setFontSize(12.0)
+        nodes[inst] = node
+        graph.addNode(node)
+    }
+
+    graph.setBgColor(Color.X11.transparent)
+    graph.setFontSize(12.0)
```
CaelmBleidd: Extract constants

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/graph/EtsCfg.kt
```diff
@@ -0,0 +1,140 @@
...
+            }
+        }
+        if (viewCatchConnections) {
+            // for (catcher in catchers(inst)) {
+            //     graph.addEdge(Edge(node.name, nodes[catcher]!!.name).also {
+            //         // it.setLabel("catch ${catcher.throwable.type}")
+            //         it.setLabel("catch")
+            //         it.setStyle(Style.Edge.dashed)
+            //     })
+            // }
```
CaelmBleidd: Why? Probably, you have to somehow connect this code with the corresponding TODO

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/model/EtsField.kt
```diff
@@ -0,0 +1,63 @@
...
+    var isProtected: Boolean = false,
+    var isReadOnly: Boolean = false,
+)
+
+// TODO
+// data class EtsDecorator(
+//     val kind: String,
+//     val content: String,
+//     val param: String,
+// )
```
CaelmBleidd: Why is it commented?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/model/EtsSignature.kt
```diff
@@ -0,0 +1,132 @@
...
+import org.jacodb.api.common.CommonMethodParameter
+import org.jacodb.panda.dynamic.ets.base.EtsType
+
+data class EtsFileSignature(
+    val projectName: String,
+    val fileName: String,
+) {
+    override fun toString(): String {
+        // Remove ".d.ts" and ".ts" file ext:
+        val tmp = fileName.replace("""(\.d\.ts|\.ts)$""".toRegex(), "")
```
CaelmBleidd: This replacement is very expensive. Why don't you use file methods or don't remove it by yourself?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/model/EtsSignature.kt
```diff
@@ -0,0 +1,132 @@
...
+}
+
+data class EtsNamespaceSignature(
+    val name: String,
+    val namespace: EtsNamespaceSignature? = null,
+    val file: EtsFileSignature? = null,
+) {
+    override fun toString(): String {
+        if (namespace != null) {
+            return "$namespace.$name"
```
CaelmBleidd: return of if expression, same in other place

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/utils/EtsFileToDot.kt
```diff
@@ -0,0 +1,137 @@
...
+                lines += "  \"${clazz.signature.name}.${etsMethod.signature.name}\" -> \"${clazz.signature.name}.${etsMethod.signature.name}.bb${etsMethod.body.cfg.blocks.first().id}.0\" [lhead=\"${clazz.signature.name}.${etsMethod.signature.name}.bb${etsMethod.body.cfg.blocks.first().id}\"];"
+            }
+
+            etsMethod.body.cfg.blocks.forEach { bb ->
+                val last = bb.stmts.lastOrNull()
+                val i = if (bb.stmts.isNotEmpty()) bb.stmts.lastIndex else 0
+                when (last ?: NopStmtDto) {
+                    is IfStmtDto -> {
+                        for ((j, succ) in bb.successors.withIndex()) {
+                            lines += "  \"${clazz.signature.name}.${etsMethod.signature.name}.bb${bb.id}.${i}\" -> \"${clazz.signature.name}.${etsMethod.signature.name}.bb${succ}.0\" [lhead=\"${clazz.signature.name}.${etsMethod.signature.name}.bb${succ}\", label=\"${if (j == 0) "true" else "false"}\"];"
```
CaelmBleidd: Can all these things be written in a more readable form?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/utils/getOperands.kt
```diff
@@ -0,0 +1,215 @@
...
+    override fun visit(stmt: EtsSwitchStmt): Sequence<EtsEntity> = sequence {
+        yield(stmt.arg)
+        yieldAll(stmt.cases)
+    }
+}
+
+private object EntityGetOperands : EtsEntity.Visitor<Sequence<EtsEntity>> {
+    override fun visit(value: EtsLocal): Sequence<EtsEntity> = sequence {
+        // empty
+    }
```
CaelmBleidd: emptySequence

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/utils/getOperands.kt
```diff
@@ -0,0 +1,215 @@
...
+
+    override fun visit(value: EtsNullConstant): Sequence<EtsEntity> = sequence {
+        // empty
+    }
+
+    override fun visit(value: EtsUndefinedConstant): Sequence<EtsEntity> = sequence {
+        // empty
+    }
+
+    override fun visit(value: EtsArrayLiteral): Sequence<EtsEntity> = sequence {
```
CaelmBleidd: Why do we even need these sequences? Why not to use lists? Isn't it too expensive?

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/utils/getOperands.kt
Lipen: That was intentional, to make the code consistent. There is no clear benefit of using `emptySequence()` here. Instead, the `sequence { /* empty */ }` is consistent and also 100% ready to be modified in any moment, for example, when copy pasting this visitor for another similar operation where those "empty visits" should be NON-empty.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/utils/EtsFileToDot.kt
Lipen: I don't think so. Writing DOT files by hand indeed can be messy. This code was not intended to be modified, it already produces good enough result (graphs are quite informative), so we can keep it as it is, for now. We can open an issue and submit another PR for "fixing", if it is really necessary. But I don't think it should block this exact PR to be merged into `neo`.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/model/EtsSignature.kt
Lipen: This was copied from ArkAnalyzer.

I do not think this operation is *very expensive*. Surely, compiling a Regex is not free, but the regex itself is simple (`$` constrains it heavily, so the matching should be fast), and the class `EtsFileSignature`, in fact, is used almost 0 times, so we can expect that this "costy" `toString` won't be a bottleneck.

Note that the straightforward `fileName.removeSuffix(".d.ts").removeSuffix(".ts")` is semantically different, since it might cut too much if the file is called `kek.ts.d.ts`. You can suggest another solution in replacement for the regex.

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/EtsTraits.kt
Lipen: It was actually a TODO.

I have fixed it to be `name == "constructor"`. We need some tests for it.

Actually, we might just drop `isConstructor` from Traits, since originally it was just a tricky adhoc, and now it seems to be unnecessary...

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/EtsTraits.kt
Lipen: This is *necessary* commented code. The "good" implementation should be "resolve the method inside cp by the signature recorded in the callexpr", however, (1) we do not have an ability to resolve methods by signature in Ets cp, and (2) callexpr (common interface) does not actually record callee signature anymore. So the "fallback" implementation is to construct the EtsMethod by hand. Note that this constructed method lacks cfg (it is uninitialized). This is bad from the design point of view. But it does not break since the taint engine does not use cfg of "callee".

Overally, it is not easy to fix it right away.

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/EtsTraits.kt
Lipen: Because someone was too lazy to think about the proper implementation.

Yes, we need it, since we have ConstantEq condition in taint configuration. We do not actually use it in taint configs for Ets, but that's another story.

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/EtsTraits.kt
Lipen: We do not need it *right now*, but it *might* be useful in the future for methods similar to `getArguments` in JcTraits. Instead of making such methods/properties to be extensions of CommonProject (for example, if your property already has some more important receiver), you either convert it to method argument (downside is that you cannot use property style anymore, and you have to pass extra arg everywhere), or you just *have* cp "in the context" -- in the Traits instance.

Current solution for `lateinit var cp: JcClasspath` in `JcTraits` should be replaced with the proper `data class JcTraits(val cp: JcClasspath)`. Similar things could be done for EtsTraits in the futute, when there will be a necessity to have cp in the context.
I am aware of it and try to fix it in nearest time.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsExpr.kt
Lipen: In fact, multidimensional arrays are not supported on the arkanalyzer side, so here we are blocked by whatever design changes they are going to perform in this regard. But yea, probably throwing an error might be okay, for now.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsExpr.kt
Lipen: There is no TernaryExpr in ArkIR, and initially, when I designed "a better version of ArkIR, in Kotlin", I though it would be nice to have it. I think we can safely drop it.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsExpr.kt
Lipen: Yes, current binary/unary operations are represented as `data class BinOp(val kind: OpEnum, args...)`. Operation's enums are purely syntactic (e.g. there are "Plus", "EqEqEq", etc), and should be replaced with a complete sealed hierarchy of different binary operations. However, for some reason, it is not the case on the ArkIR side - they also represent binops as op+args (even worse, they have "op = string").

I am also aware of this issue and try to provide these sealed hierarchy in some time.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/Visitors.kt
Lipen: The whole file "Visitors" was just for testing.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Model.kt
Lipen: No, we have `@JsonClassDiscriminator("_")` annotations on all DTOs. I was not sure whether they work correctly, so I kept `classDiscriminator = "_"` setting, as we commonly used it before annotations. The annotations seem to be working correctly, so we can drop these commented settings.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Model.kt
Lipen: Obsolete

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Values.kt
Lipen: Currently constants in ArkIR are represented just as `Constant(value: string, type: Type)`. Note that ALL constants are stored as strings. That's wild.

One day, ArkIR is going to have proper constants (sealed interface Constant with different inheritors such as `StringConstant`, `NumberConstant`, etc). And we will be ready for this moment with this commented code. In fact, we should probably do this redesign in ArkIR ourselves, and submit PR to them with similar classes.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Values.kt
Lipen: Objects and arrays were (and probably are still) a mess in ArkIR. Commented code is the expected "good solutions". But it might not match ArkIR at all. We should re-check it.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/utils/getOperands.kt
Lipen: In fact, it does not matter. The performance implications will be negligible here. However, theoretically, expressions might be very deeply nested, so I personally prefer lazy solutions (i.e., Sequence) in such cases.

jacodb-analysis/src/main/kotlin/org/jacodb/analysis/util/EtsTraits.kt
CaelmBleidd: In this case, you should create a todo (better with a linked issue, but in our case with neo branch that's not easy) with a comment about it. Otherwise, a commented code will be surely lost at some point

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsExpr.kt
CaelmBleidd: Is there an issue about it in their repository? Please, check if it is still a thing and create an issue about it

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/utils/EtsFileToDot.kt
CaelmBleidd: I agree, but, for example, replacing long access instructions with variables shouldn't cause too much pain, and it already will look better

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/Values.kt
Lipen: Turns out, objects are *much* more complex than just "a map with keys and values". There are 5 different "property element kinds" (that is, things that might be inside curly braces of object literals): property assignment (`let obj = { a: 42 }`), shorthand property assignment (`let obj = { a }`), spread assignment (`let obj = { ...other }`), method declaration (`let obj = { method() {...} }`), accessor declaration (`let obj = { get accessor() {...} }`, also `set`). All of them are syntactic constructions and cannot be "simplified" into a list of 'key:value' pairs, since the final result can only be determined at the runtime (due to spread operator, for example). Also note that the order of "property-elements" is important (again, due to spread operator).

I think we can design it that way on our side, but ultimately we have to wait for the support in ArkIR on the arkanalyzer side, and them adapt to it. Or force it by ourselves via PRs there.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/EtsExpr.kt
Lipen: As far as I understood after some research, "native" multi-dimensional arrays do not exist in JS/TS -- there are only nested arrays. So, `number[][]` is just `(number[])[]`, or rather, `Array<Array<number>>`. That is, ArkIR won't "support" MDA, neither will we. And the current design of arrays is OK-ish enough.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/base/Ops.kt
Lipen: Note that the original "operators" are just strings. During the process of converting DTOs from ArkIR (TS) into EtsIR (Kotlin), we accept only the listed operators, and transform them into enum variants. This is just a syntactic operation, we do not rely on any semantic of these operators. The original operators were arbitrary strings, so the responsibility of matching the semantics with these string was on the shoulders of the user. The same flavor was kept in our implementation - we provide a enum with AST-like operators (note that I have extracted these names from `swc` JS/TS parser) - and the user will make a decision on what "EqEqEq" mean for him.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
Lipen: All files that contain classes are spelled in TitleCase. All files that only contain utility functions, or even just one function, better be spelled the same as functions are spelled - in camelCase, don't you think so?

Feel free to ping me once again and I am going to change it for you. Or feel free to rename them yourself :)

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
Lipen: I am not sure whether we even need it, or how to "properly" implement it (hence the *triple* question mark in the comment, that's rare). If we override equals (delegate to signature comparison), then different "unknown statements" will get lost after you put them in a set. That might be a good behavior for "normal" statements, but "unknown statements" are indicators of "something is incomplete/wrong", so we probably do not want to lose them in sets. The default ref-equals seems to be OK for them.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
Lipen: This is a single code block.

jacodb-panda-dynamic/src/main/kotlin/org/jacodb/panda/dynamic/ets/dto/convert.kt
Lipen: What's the difference here? `!!` is more concise. And we guarantee it cannot fail here, just by the consturction.

Lipen (approved):
Perfect code, outstanding developers, bug-free masterpiece.

jacodb-analysis/src/test/kotlin/org/jacodb/analysis/impl/IfdsUnusedTest.kt
```diff
@@ -57,7 +57,7 @@ abstract class IfdsUnusedTest : BaseAnalysisTest() {
         )
     }
 
-    @Disabled
+    @Disabled("#220")
```
CaelmBleidd, line 60: What is 220?

Lipen: Issue number
https://github.com/UnitTestBot/jacodb/issues/220

PR #236, opened 2024-07-12
Ir Fixes

Saloed:
In BaseInstructionsTest, avoid doing  flow graph checking for methods having "StringConcat$" substring in their names in addition to "$$forInline" and "lambda$".

[jacodb-core] Fix minor review points

[build] Upgrade org.ow2.asm dependencies to version 9.7

[jacodb-core] IR: allow lambdas have line number equal to 0

[jacodb-core] IR: fix frame preparation for label node having several unprocessed predecessors

If a label instruction node has several unprocessed predecessors (say, jump instruction nodes), their frames are all equal to `null`. Process this case separately on preparation of frame for the label instruction node.


CaelmBleidd (approved):
Lgtm

PR #243, opened 2024-07-24
Set superclass field for EtsClass

zishkaz:


jacodb-ets/src/main/kotlin/org/jacodb/ets/model/EtsSignature.kt
```diff
@@ -65,6 +65,10 @@ data class EtsClassSignature(
             name
         }
     }
+
+    override fun equals(other: Any?): Boolean = (other as? EtsClassSignature)?.let { sig ->
```
Lipen: @zishkaz Please, rewrite it in a common way, as block function, do not squeeze everything into one line. Also, add some TODO/comment, explaining why we need such `equals` implementation, and how/when it will be changed back to the default data class' equals.

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/EtsClass.kt
```diff
@@ -33,7 +33,7 @@ class EtsClassImpl(
     override val methods: List<EtsMethod>,
     override val ctor: EtsMethod,
 ) : EtsClass {
-    override var superClass: EtsClass? = null
+    override var superClass: EtsClassSignature? = null
```
Lipen: Since `superClass` is now just a signature, we can move it in the constructor and make immutable.

jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Convert.kt
```diff
@@ -430,6 +430,14 @@ fun convertToEtsClass(classDto: ClassDto): EtsClass {
         file = null, // TODO
     )
 
+    val superclassSignature = classDto.superClassName?.takeIf { it != "" }?.let { spName ->
```
Lipen, line 433: Why `spName`? Maybe just `name`?

Lipen (changes requested):

jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Convert.kt
zishkaz: Not to overlap with potential "name" variable in a higher scope

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/EtsClass.kt
```diff
@@ -18,7 +18,7 @@ package org.jacodb.ets.model
 
 interface EtsClass {
     val signature: EtsClassSignature
-    val superClass: EtsClass?
+    val superClass: EtsClassSignature?
```
Lipen: Move it to the end (before `name`, as `name` is a re-export).

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/EtsClass.kt
```diff
@@ -32,8 +32,8 @@ class EtsClassImpl(
     override val fields: List<EtsField>,
     override val methods: List<EtsMethod>,
     override val ctor: EtsMethod,
+    override val superClass: EtsClassSignature?,
```
Lipen: Maybe make it `null` by default?

Lipen (changes requested):

Lipen (approved):
Nice!

PR #249, opened 2024-08-02
Refactor Ets values, operations, visitors

Lipen:
This PR introduces some improvements in the design of Ets operations (enums are replaced with classes) and also simplifies `EtsValue` visitors (merge visitors for Immediate, Constant, Ref into one).

In addition, the type of `EtsIfStmt.condition` was changed from `EtsConditionExpr` to `EtsEntity`, since JS/TS allows "implicit" constructions such as `if (x) {...}`. See also: https://gitee.com/openharmony-sig/arkanalyzer/issues/IAHIJ5


jacodb-ets/src/main/kotlin/org/jacodb/ets/base/Ops.kt
```diff
@@ -16,260 +16,41 @@
...
-    NullishAssign,
+object Ops {
+    const val MINUS = "-"
+    const val PLUS = "+"
+    const val BANG = "!"
+    const val TILDE = "~"
+    const val PLUS_PLUS = "++"
+    const val MINUS_MINUS = "--"
+    const val EQ_EQ = "=="
+    const val NOT_EQ_EQ = "!="
```
CaelmBleidd: why do we need this additional EQ?

CaelmBleidd (approved):

jacodb-ets/src/main/kotlin/org/jacodb/ets/base/Ops.kt
Lipen: That's typo. I am gonna fix this and several other "bugs" you have not noticed later today.

PR #251, opened 2024-08-05
Refactor DOT dump for Ets files

rudolf101:


jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/Dot.kt
```diff
@@ -0,0 +1,41 @@
...
+
+package org.jacodb.ets.utils
+
+import mu.KotlinLogging
+import java.io.File
+import kotlin.time.Duration.Companion.seconds
+
+private val logger = KotlinLogging.logger {}
+
+fun render(path: String, dump: (File) -> Unit) {
```
Lipen: The part `dump(dotFile)` does not belong to "render". I would suggest removing from here and perform it outside. "Render" function should only "render DOT to PDF", preferably _in the background_, though the latter might be a little bit harder to implement: you could return job (thread/process) handle and await (all of them, if doing rendering in batch) outside.

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/EtsFileDtoToDot.kt
```diff
@@ -182,6 +183,37 @@ fun EtsFileDto.toDot(useLR: Boolean = false): String {
     return lines.joinToString("\n")
 }
 
+fun EtsFileDto.dumpHuimpleTo(output: BufferedWriter) {
```
Lipen: You can also use `with(output)` so that all `output.writeln` calls become just `writeln` in the context.

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/Dot.kt
```diff
@@ -0,0 +1,41 @@
...
+    val dotFile = File(path)
+    dump(dotFile)
+    logger.info { "Generated DOT file: ${dotFile.absolutePath}" }
+    for (format in listOf("pdf")) {
+        val formatFile = dotFile.resolveSibling(dotFile.nameWithoutExtension + ".$format")
```
Lipen: We will probably also need the ability to specify the output path (for PDF) instead of it being a sibling of a DOT file.
Also, you can add an argument for a list of formats, `listOf("pdf")` by default.

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/EtsFileDtoToDot.kt
```diff
@@ -182,6 +183,37 @@ fun EtsFileDto.toDot(useLR: Boolean = false): String {
...
+        output.writeln("  methods: ${clazz.methods.size}")
+        clazz.methods.forEach { method ->
+            output.writeln("  - METHOD '${method.signature}':")
+            output.writeln("    locals = ${method.body.locals}")
+            output.writeln("    typeParameters = ${method.typeParameters}")
+            output.writeln("    blocks: ${method.body.cfg.blocks.size}")
+            method.body.cfg.blocks.forEach { block ->
+                output.writeln("    - BLOCK ${block.id} with ${block.stmts.size} statements:")
+                block.stmts.forEachIndexed { i, inst ->
+                    output.writeln("      ${i + 1}. $inst")
```
Lipen: We also need successors, either for blocks (in DTOs?) or for statements (in Ets model?).

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/EtsFileDtoToDot.kt
Lipen: This function does not belong to `EtsFileDtoToDot`. Either rename the file, or move the function to a separate file (or to Utils).

jacodb-ets/src/test/kotlin/org/jacodb/ets/test/EtsFromJsonTest.kt
```diff
@@ -72,13 +73,22 @@ class EtsFromJsonTest {
...
         val etsFile = loadEtsFileAutoConvert(res.path)
         println("etsFile = $etsFile")
     }
 
+    @Test
+    fun testLoadEtsFileAutoConvertWithDot() {
+        val path = "/source/example.ts"
+        val res = this::class.java.getResource(path)
+            ?: error("Resource not found: $path")
+        val etsFile = loadEtsFileAutoConvertWithDot(res.path, res.path.substringBeforeLast("/"))
```
Lipen: Note that the path `res.path` might NOT be the one you have expected. "Resource" path is NOT `src/test/resources`, but `build/resources/test`!
It is better to specify a temporary folder here, for temporary DOTs in tests.

jacodb-ets/src/test/kotlin/org/jacodb/ets/test/utils/Entrypoints.kt
```diff
@@ -0,0 +1,133 @@
...
+            val outputDir = dotDir.resolve(path.parent ?: dotDir)
+            if (!outputDir.exists()) {
+                outputDir.createDirectories()
+            }
+            val outputPath = outputDir.resolve("${name}${suffix}.dot").toString()
+            render(outputPath) { file ->
+                when (fileData) {
+                    is EtsFileDto -> fileData.dumpDot(file)
+                    is EtsFile -> fileData.dumpDot(file)
+                }
```
Lipen: That's unfortunate.

jacodb-ets/src/test/kotlin/org/jacodb/ets/test/utils/Entrypoints.kt
```diff
@@ -0,0 +1,133 @@
...
+            }
+        } catch (e: Exception) {
+            logger.error { e.stackTraceToString() }
```
Lipen: These entrypoints are to be called manually anyway. Why not just let them unwind on exception (i.e. remove try/catch here) instead of semi-silently ignoring them? Imagine the situation: you enter all the necessary paths to the files you need to be rendered, hit Run, await, receive green checkmark in IDEA, go open the resulting PDFs. Most probably, you just don't notice than an exception has occurred.

Lipen (changes requested):

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/EtsFileToDot.kt
```diff
@@ -146,6 +147,32 @@ fun EtsFile.toDot(useLR: Boolean = false): String {
     return lines.joinToString("\n")
 }
 
+fun EtsFile.dumpHuimpleTo(output: BufferedWriter) {
```
Lipen: `dumpContentTo`

Lipen (approved):
Nice!

PR #253, opened 2024-08-05
Fix locations in intermediate statements 

Lipen:
This PR fixes some spurious locations mismatches by extracting the recursive calls (such as `EtsAssignStmt(location = loc(), lhv = <recurse>, rhv = <recurse>)`) with `ensureOneAddress` inside them. Without extracting these sub-expressions, `loc()` produces a location that might be "reused" by another `loc()` call inside `ensureOneAddress` (which creates extra assignment statement for complex expressions, thus allocates and fills a location). The proper way of doing it is to ensure that `lhv` and `rhv` are computed (and all necessary extra statements are already created and put inside `currentStmts`) *before* calling `loc()` for the newly created statement.
In fact, not all extractions that were made in this PR are strictly necessary for the correctness, but I just made the code changes uniform and consistent.

### DIFF
--- a/jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Convert.kt
+++ b/jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Convert.kt
@@ -180,48 +180,72 @@ class EtsMethodBuilder(
                 }
             }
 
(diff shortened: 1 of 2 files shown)


CaelmBleidd (approved):
LGTM

PR #255, opened 2024-08-06
Update code to match the current ArkAnalyzer

Lipen:
This PR introduces some changes that are necessary to upgrade to the current version of ArkAnalyzer.

- Bump ArkAnalyzer, switch to using a special neo-branch in the fork.
- Update JSON dumps of ArkIR for samples.
- Remove fields initializers. They were moved (as a list of assignments: `this.field := value`) into the special `@instance_init` method.
- Make `MethodDto.body` nullable since it might be absent.
- Support AwaitExpr and YieldExpr.
- Rename CallableType to FunctionType.
- Add JsonElement inside UnknownValueDto.
- Add empty CFG. EtsMethod returns empty CFG when it does not have one.
- Use kotlin("test") for Kotlin-style assertions in tests.


rudolf101 (approved):
🌟 This code shines brighter than my future. Ready for merge!

PR #256, opened 2024-08-07
IR fixes and features improvement

Saloed:
[jacodb-core] Minor: remove unnecessary .toList(), fix typo

[jacodb-core] Make Builders working correctly with same classes from different locations

[jacodb-core] Migrate to Kotlin 1.9.25 and kotlinx.metadata 0.9.0

[jacodb-core] RawInstListBuilder: less strict requirement of types compatibility

When copying locals for a new frame, consider argument of compatible type as a suitable variable. Types are compatible if they both are reference types, or both are primitive types of the same size (32 bits or 64 bits).

[jacodb-core] RawInstListBuilder: never copy argument of different type into new frame

[jacodb-core] RawInstListBuilder: never copy Kotlin's implicit argument "it" into new frame

Inline methods compiled with Kotlin can use single name ("it") for JcRawArguments having different types in different frames. So inheriting of such local from an outer frame can result in an invalid operation.

[jacodb-core] RawInstListBuilder: define two-pass calculation of try-catch handlers' predecessors

This commit changes the way how predecessors of try-catch blocks' handlers are calculated. In some specific cases, there can be several try-catch blocks with single handler. In that case, in order to get rid of the handler predecessors not belonging to all its try-catch blocks, we have to filter them in a separate pass though all try-cache blocks.

[jacodb-core] RawInstListBuilder: for try-catch handlers, use only outside the block predecessors

[jacodb-core] RawInstListBuilder: simplify calculation of predecessor frames for try-catch handlers

Get rid of traversing graph of instructions on calculation of predecessor frames for try-catch handlers. Instead, get correct predecessors of try-catch handlers (outside their blocks) in the buildGraph() method.

[jacodb-core] RawInstListBuilder: extract extension tryCatchBlocks() for a LabelNode

[jacodb-core] RawInstListBuilder: for try-catch handlers, use predecessors' frames outside the block

For a try-catch handler, we should not take predecessor frames from try-catch block itself, but we should try to find suitable predecessor frames outside the block.

[jacodb-core] RawInstListBuilder: correct suitable predecessors' frames for TryCatchBlockNodes handlers

Generally, any instruction in a TryCatchBlockNodes handler can use locals declared outside the block. However, prior to this commit, all predecessors of the handler were instructions from the block only. So for a TryCatchBlockNode handler, we should calculate suitable predecessor frames installed outside the block in addition to those calculated by predecessors from the block.

[jacodb-core] RawInstListBuilder: minor optimization of building CFG for TryCatchBlockNodes

[jacodb-core] RawInstListBuilder: get rid of "dead" instructions, update graph correspondingly

There are no longer frames not connected to the man CFG, so we no longer have a need in "dead instructions". Each instruction has a non-empty list of predecessors.

[jacodb-core] RawInstListBuilder: handle correctly stacks of "dead" frames

There can be some frames not connected to the man CFG. As of now, they are processed correctly - without errors and exceptions. But TODO: completely remove their instructions from instruction list.

[jacodb-core] RawInstListBuilder: never override later stack assignments

This change is similar to that for later assignments. It would probably fix some unknown bugs.

[build] Use ASM 9.6

[jacodb-core] RawInstListBuilder: on building frame, avoid LabelNodes being put to dead instructions

For FrameNode, search for LabelNode not being put to dead instructions and use this node for getting block predecessors.

[jacodb-core] RawInstListBuilder: never override later assignments

For JumpInsnNode, there can be several unknown frames at a time. In that case, later assignments for variables of some frame could clear those for another. To address this, collect later oin a flat list, so that later assignments would never be cleared.

[jacodb-core] RawInstListBuilder: simplify isBetween()

[jacodb-core] RawInstListBuilder: for non-static method, always make sure that first local variable is `JcRawThis`

[jacodb-core] Refactor copyLocals() in order to make it more readable and debuggable

[jacodb-core] In InMemoryHierarchy.syncQuery(), return lazily built sequence

[jacodb-core] In HierarchyExtensionERS.explicitSubClasses(), return lazily built sequence

[jacodb-core] In HierarchyExtensionSQL, do not use JCDBContext lazily in BatchedSequence lambda

### DIFF
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -3,7 +3,7 @@
 import org.gradle.plugin.use.PluginDependenciesSpec
 
 object Versions {
-    const val asm = "9.7"
+    const val asm = "9.6"
     const val dokka = "1.7.20"
     const val gradle_download = "5.3.0"
     const val gradle_versions = "0.47.0"
@@ -22,13 +22,13 @@ object Versions {
     const val jooq = "3.14.16"
     const val juliet = "1.3.2"
     const val junit = "5.9.2"
-    const val kotlin = "1.7.21"
+    const val kotlin = "1.9.25"
     const val kotlin_logging = "1.8.3"
     const val kotlinx_benchmark = "0.4.6"
     const val kotlinx_cli = "0.3.5"
     const val kotlinx_collections_immutable = "0.3.5"
     const val kotlinx_coroutines = "1.6.4"
-    const val kotlinx_metadata = "0.5.0"
+    const val kotlinx_metadata = "0.9.0"
     const val kotlinx_serialization = "1.4.1"
     const val licenser = "0.6.1"
     const val mockk = "1.13.3"
--- a/jacodb-core/src/main/kotlin/org/jacodb/impl/JcDatabaseImpl.kt
+++ b/jacodb-core/src/main/kotlin/org/jacodb/impl/JcDatabaseImpl.kt
@@ -23,7 +23,7 @@ import org.jacodb.impl.features.classpaths.KotlinMetadata
 import org.jacodb.impl.features.classpaths.MethodInstructionsFeature
 import org.jacodb.impl.fs.JavaRuntime
 import org.jacodb.impl.fs.asByteCodeLocation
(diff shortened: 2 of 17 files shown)


CaelmBleidd (approved):
LGTM

PR #258, opened 2024-08-08
Generate local test resources via Gradle task

Lipen:
This PR proposes to start relying on local test resources (JSON dumps of EtsIR, etc) that can be generated via `generateTestResource` Gradle task.

In addition, this PR includes some updates for DOT generation and PDF rendering for samples (all local test resources), see `DumpEtsFilesToDot` in `Entrypoints.kt`.


jacodb-ets/build.gradle.kts
```diff
@@ -21,32 +21,46 @@ dependencies {
...
     description = "Generates test resources from TypeScript files using ArkAnalyzer."
     doLast {
+        println("Generating test resources using ArkAnalyzer...")
+        val startTime = System.currentTimeMillis()
+
         val envVarName = "ARKANALYZER_DIR"
-        val defaultArkAnalyzerDir = "../arkanalyzer"
+        val defaultArkAnalyzerDir = "arkanalyzer"
 
         val arkAnalyzerDir = rootDir.resolve(System.getenv(envVarName) ?: run {
```
rudolf101, line 37: Should we use `rootDir.resolveSibling` if the system environment variable is not set? Either way, it won't affect the path resolving if the environment variable is set.

rudolf101 (changes requested):

rudolf101 (approved):
💻 If only all code were this smooth. Merging now!

PR #263, opened 2024-08-16
Improve resolution of callees in `EtsApplicationGraph`

Lipen:
This PR improves the resolution of callees in `EtsApplicationGraph`. In a perfect world, matching suitable methods by their signature would be enough, however in some cases the available signature is incomplete, e.g. enclosing class might be missing. In order to still be able to resolve such calls, we consequently try to resolve the (single) callee: first, by name only, and next, by name and class. If these lookups fail (for example, if there are still multiple methods with matching names and enclosing classes, which be in the case of a multi-file project), we simply fall-back to the default behavior and match by the whole signature. Overall, we still rely on the signature to uniquely represent a method.


rudolf101 (approved):

Lipen:
This PR is not necessary anymore, sorry.

PR #280, opened 2024-12-05
Remove jacodb-analysis module

Lipen:
This PR removes `jacodb-analysis` module, which was moved into [USVM](https://github.com/UnitTestBot/usvm) (see https://github.com/UnitTestBot/usvm/tree/main/usvm-dataflow).

Note that this PR also removes some tests in `jacodb-ets` (related to analysis, they are already being moved to USVM), and modules `jacodb-panda-static` (obsolete) and `jacodb-cli` (heavily depend on `jacodb-analysis`). FWIW, it should not be a problem, as these modules are barely used by anyone, but the README and documentation now becomes somewhat outdated.


lehvolk:
Let's call it Volkswagen PR at the moment

Saloed:
We have some analysis missed from the USVM. Since it is jvm specific we can move it to `usvm-jvm-dataflow`. 

https://github.com/UnitTestBot/jacodb/blob/neo/jacodb-analysis/src/main/kotlin/org/jacodb/analysis/impl/custom/ReachingDefinitionsAnalysis.kt

Lipen:
@Saloed `ReachingDefinitionsAnalysis` is unused in jacodb, should we still keep it, even so there are no tests for it?
As for the `NullAssumptionAnalysis`, it has some tests in jacodb, but these tests depend on `BaseTest`, which is defined in `jacodb-core`'s `testFixture` module, which is not published, so these tests in `jacodb-analysis/test` cannot be simply transferred to USVM, and without the tests there is not point in transferring only the analyses. Should we nevertheless move the whole test infrastructure for these analyses to USVM? Or maybe make the `testFixture` module in `jacodb-core` publishable?

Lipen:
https://github.com/UnitTestBot/usvm/pull/222 was merged, all analyses from `jacodb-analysis` are now in `usvm-dataflow`

Saloed (approved):

PR #313, opened 2025-04-14
Change ETS model

Lipen:
This PR introduces some changes in ETS IR model.


jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Convert.kt
```diff
@@ -16,156 +16,151 @@
...
-    private fun newTempLocal(type: EtsType): EtsLocal {
-        return EtsLocal("_tmp${freeTempLocal++}", type)
+    private fun newTempLocal(): EtsLocal {
+        return EtsLocal("_tmp${freeTempLocal++}")
     }
 
-    private fun loc(): EtsInstLocation {
-        return EtsInstLocation(etsMethod, currentStmts.size)
+    private fun loc(): EtsStmtLocation {
+        return EtsStmtLocation.stub(method)
```
CaelmBleidd, line 146: Why we don't save location indices anymore?

jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Convert.kt
```diff
@@ -415,137 +396,75 @@ class EtsMethodBuilder(
...
-            body = body,
-        )
-    }
-
     val signature = signature.toEtsClassSignature()
     val superClassSignature = superClassName?.takeIf { it != "" }?.let { name ->
         EtsClassSignature(
             name = name,
-            file = EtsFileSignature.DEFAULT,
+            file = EtsFileSignature.UNKNOWN,
```
CaelmBleidd, line 448: Why? Do we never know a signature for a file?

jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Ops.kt
```diff
@@ -45,7 +45,7 @@ object Ops {
         const val AND = "&&"
         const val OR = "||"
         const val NULLISH = "??"
-        const val COMMA = ","
+        // const val COMMA = ","
```
CaelmBleidd, line 48: Why?

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/CfgBlock.kt
```diff
@@ -0,0 +1,95 @@
...
+        get() = linear.entries
+    override val exits: List<EtsStmt>
+        get() = linear.exits
+
+    override fun successors(stmt: EtsStmt): Set<EtsStmt> = linear.successors(stmt)
+    override fun predecessors(stmt: EtsStmt): Set<EtsStmt> = linear.predecessors(stmt)
+    override fun throwers(node: EtsStmt): Set<EtsStmt> = linear.throwers(node)
+    override fun catchers(node: EtsStmt): Set<EtsStmt> = linear.catchers(node)
+
+    // val stmts: List<EtsStmt> by lazy {
```
CaelmBleidd: Lets remove it

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Entity.kt
```diff
@@ -14,11 +14,13 @@
...
  */
 
-package org.jacodb.ets.base
+package org.jacodb.ets.model
 
 import org.jacodb.api.common.cfg.CommonExpr
 
 interface EtsEntity : CommonExpr {
+
+    @Deprecated("Do not rely on it!")
```
CaelmBleidd: Please, clarify the reason

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Entity.kt
```diff
@@ -52,10 +54,12 @@ interface EtsEntity : CommonExpr {
 data class EtsRawEntity(
     val kind: String,
     val extra: Map<String, Any> = emptyMap(),
-    override val type: EtsType,
 ) : EtsEntity {
+    override val type: EtsType
```
CaelmBleidd, line 57: "This declaration overrides a deprecated member but is not marked as deprecated itself. Please add the '@Deprecated' annotation or suppress the"

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Ref.kt
```diff
@@ -14,17 +14,20 @@
  *  limitations under the License.
  */
 
-package org.jacodb.ets.base
+@file:Suppress("OVERRIDE_DEPRECATION")
```
CaelmBleidd: Why do you suppress it instead of marking as deprecated?

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/StmtLocation.kt
```diff
@@ -0,0 +1,49 @@
...
+    var index: Int,
+) : CommonInstLocation {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as EtsStmtLocation
+
+        if (method != other.method) return false
+        // if (index == -1 || other.index == -1) return false
```
CaelmBleidd: Why is it commented?

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/BlockCfgBuilder.kt
```diff
@@ -0,0 +1,292 @@
...
+            type = EtsUnknownType, // TODO
+        )
+
+        is Parameter -> EtsParameterRef(
+            index = index,
+            type = EtsUnknownType, // TODO
+        )
+
+        ThisRef -> EtsThis(
+            type = EtsUnknownType, // TODO
```
CaelmBleidd, line 171: Why todos?

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/Handler.kt
```diff
@@ -16,85 +16,87 @@
...
+
+abstract class AbstractHandler : EtsEntity.Visitor<Unit>, EtsStmt.Visitor<Unit> {
 
     abstract fun handle(value: EtsEntity)
     abstract fun handle(stmt: EtsStmt)
 
-    final override fun defaultVisit(value: EtsEntity) {
-        handle(value)
-    }
+    // final override fun defaultVisit(value: EtsEntity) {
```
CaelmBleidd: Lets remove it?

CaelmBleidd (changes requested):

jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Convert.kt
Lipen: Of the superclass? Yes, it is not resolved yet, we only known the "name" of the superclass.

jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Ops.kt
Lipen: There is no `CommaExpr` in ArkIR, and we not have `EtsCommaExpr` in EtsIR, so we do not need to support COMMA token in dtos -- it would be an error if we encounter it.

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/StmtLocation.kt
Lipen: Because I needed it when I first wrote (during the CFG refactor) it for some reason, but lately I have discovered that this leads to a bug (NopStmt at -1 position are indistinguishable, thus cannot be used as keys in maps, or something like that), so I have just commented it for now. Probably need to discuss whether it is still necessary, and if it is, then how to avoid the aforementioned bug.

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/BlockCfgBuilder.kt
Lipen: Because we probably _can_ pass some outer class (enclosing the method for which for build the CFG) and use it as a type for ThisRef.

jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Convert.kt
Lipen: Because this `loc` is used when we convert THEIR block cfg to OUR block cfg. Statements in block cfg yet do not have location indices. They only get them when we linearize the cfg.

Before the whole CFG refactor, we built our linear CFG directly from their Block CFG DTO, thus we had `index = currentStmts.size` as the location index.

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Entity.kt
Lipen: Literally, DO NOT RELY ON THIS. THE TYPE IS A LIE.

Lipen: Currently, this deprecation is just to clearly see (since it emits warnings which prohibit usvm from compiling, for example) all the places where we use `type` property. In all of them, we either explicitly suppress it, or somehow stop using it completely.

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Entity.kt
Lipen: Probably have missed file-wise suppression in this file...

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/Handler.kt
Lipen: Probably yes, I have decided to remove default visitors here because it is hard to introduce new inheritors and not forget to include the corresponding visit in all visitors that we provide - the compiler does not help to spot such places when we have impl `Default` visitor interface.

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/CfgBlock.kt
Lipen: It is to be discussed, whether we should rely on linear cfg to provide a list of statements (thus _triggerring the lazy construction_ of the linear cfg itself), or we can simply get this list directly from block cfg via manual traversal here.

Relying on the linear cfg is more straightforward and avoids code duplication.

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Ref.kt
Lipen: :panda_face:

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Entity.kt
CaelmBleidd: It might seem obvious for you, but it isn't in real life. Maybe I cannot call it because it might fail, of because calling it will take too much time, or because want to force using of another API. There might be many reasons and users should not guess

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/BlockCfgBuilder.kt
CaelmBleidd: But seems like we skip all the types since they all are marked as TODO. At the same time, somewhere else I've seen usage of this type, so we have logic that might take some advantage out of this information

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Entity.kt
Lipen: Removed the deprecation annotation

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/BlockCfgBuilder.kt
Lipen: Actually, this is just a DSL CFG, so probably we even do not need to solve its problems. DSL for programs lacks many typing information.

Lipen: Actually, this CFG builder is for converting from "DSL CFG" to "ETS CFG". DSL for programs lacks many typing information, and this is intentional and pretty much expected, so probably we do not need to try fixing all its problems.

CaelmBleidd (approved):

PR #318, opened 2025-05-05
Add Protobuf/gRPC-based EtsIR loader

Lipen:
This PR introduces a Protobuf-based EtsIR loader that communicates with the local ArkAnalyzer gRPC server, replacing the previous JSON-based approach.

Currently, this PR contains two different implementations of Protobuf models of EtsIR: auto-generated (via Gradle plugin that wraps `protoc`) and manual-written (with `kotlinx-serialization-protobuf`).
- We have to decide which one is more convenient for us.
- We also need to take a look at [Wire](https://square.github.io/wire/) --- it can auto-generate Kotlin classes (which are _much_ better than Java classes from `protoc`) from `.proto` models.
- After selecting the desired final solution, we are going to clean up this PR.

In addition, this PR contains the implementation of an HTTP client (using [Ktor](https://github.com/ktorio/ktor)) for sending requests to a HTTP ArkAnalyzer server, internally using either JSON or ProtoBuf format. This shows the benefits of having the proto-model written with kotlinx-serialization, and not standalone auto-generated files. However, currently there is no simple way of easily switching between HTTP and gRPC protocols, so by using Ktor as a client, we are limited to HTTP-based solutions --- most importantly, from the ArkAnalyzer side as well, where we already began integrating gRPC server (after discussing that gRPC is indeed more suitable and future-proof for our means). Thus, we probably end up deleting all Ktor-related code from this PR, and the kx-serialization model as well, since it can't be (easily) used in auto-generated gRPC client/server stubs instead of the auto-generated messages.

Overall, we have two possible final solutions:
- Official protobuf and grpc tools. Auto-generated (`protoc`) Java classes for model, Kotlin type-safe wrappers, auto-generated gRPC client/server stubs. Works with Java 8.
- Wire. Similar to the first solution, but _different_. Also provides auto-generated model (pretty convenient Kotlin classes via kotlin-poet), client/server stubs. Requires Java 11.

Note: the description above mentions both client and server, though in this particular PR we only have to introduce a client for the ArkAnalyzer gRPC server. However, the same arguments will be applicable when we are going to develop the USVM server for the type inference. This PR is just a playground for all these possibilities.


.github/workflows/build-and-test.yml
```diff
@@ -64,7 +62,8 @@ jobs:
           DEST_DIR="arkanalyzer"
           MAX_RETRIES=10
           RETRY_DELAY=3  # Delay between retries in seconds  
-          BRANCH="neo/2025-04-14"
+          #BRANCH="neo/2025-04-14"
+          BRANCH="lipen/grpc"
```
CaelmBleidd: Replace before merge

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -339,6 +342,99 @@ object Libs {
...
+        name = "ktor-server-core",
+        version = Versions.ktor
+    )
+    val ktor_server_netty = dep(
+        group = "io.ktor",
+        name = "ktor-server-netty",
+        version = Versions.ktor
+    )
+    val ktor_server_content_negotiation = dep(
+        group = "io.ktor",
```
CaelmBleidd: Actualize dependencies

CaelmBleidd (approved):

jacodb-ets/src/main/kotlin/org/jacodb/ets/grpc/LoadScene.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlin.time.measureTimedValue
+import model.Scene as ProtoScene
+
+private val logger = KotlinLogging.logger {}
+
+const val DEFAULT_PORT = 50051
+
+fun loadScene(
+    path: Path,
+    port: Int = DEFAULT_PORT,
```
Lipen: Seems like the existence of the default value for `port` argument leads to user errors due to the implicit dependency on "externally started" AA server on this default port 50051. It is the user's responsibility to start the server and provide us the port to try to load the Scene from.

TODO: remove `DEFAULT_PORT`.

jacodb-ets/src/main/kotlin/org/jacodb/ets/grpc/LoadScene.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlin.time.DurationUnit
+import kotlin.time.measureTimedValue
+import model.Scene as ProtoScene
+
+private val logger = KotlinLogging.logger {}
+
+const val DEFAULT_PORT = 50051
+
+fun loadScene(
+    path: Path,
```
Lipen, line 33: TODO: add `inferTypes: Int? = null` and fill the request with this parameter

PR #330, opened 2025-07-16
Make raw serializers more lenient

Lipen:
This PR makes the serializers for `EtsRawStmt`, `EtsRawValue`, `EtsRawType` more lenient, i.e., allow them accept JSONs that use `kind` discriminator instead of `_`.


jacodb-ets/src/main/kotlin/org/jacodb/ets/dto/Serializers.kt
```diff
@@ -91,10 +91,17 @@ object RawStmtSerializer : KSerializer<RawStmtDto> {
     override fun deserialize(decoder: Decoder): RawStmtDto {
         require(decoder is JsonDecoder)
         val element = decoder.decodeJsonElement().jsonObject
-        val kind = element.getValue("_").jsonPrimitive.content
-        val details = element.toMutableMap()
-        details.remove("_")
-        return RawStmtDto(kind, JsonObject(details))
+        if ("_" in element.keys ) {
```
Lipen:
```suggestion
        if ("_" in element.keys) {
```

CaelmBleidd (approved):

PR #332, opened 2025-07-22
Added versions for approximations

MchKosticyn:


Saloed (approved):

Lipen:
@Saloed USVM needs an update after this PR, since `usvm-jvm` depends on the modified classes.

Currently, USVM does not compile with the latest jacodb.

PR #340, opened 2025-08-06
Improve TS CFG DSL

Lipen:
This PR improves CFG DSL by adding boolean/integer/string constants and `CustomValue` (for arbitrary `EtsEntity`), and also by refining if/elseif/else stmt builders.

### DIFF
--- a/jacodb-ets/src/main/kotlin/org/jacodb/ets/dsl/ProgramBuilder.kt
+++ b/jacodb-ets/src/main/kotlin/org/jacodb/ets/dsl/ProgramBuilder.kt
@@ -18,35 +18,60 @@
 
 package org.jacodb.ets.dsl
 
+import org.jacodb.ets.model.EtsEntity
+
 interface ProgramBuilder {
+    fun nop()
     fun assign(target: Local, expr: Expr)
     fun ret(expr: Expr)
-    fun ifStmt(condition: Expr, block: IfBuilder.() -> Unit)
-    fun nop()
     fun label(name: String)
     fun goto(label: String)
+    fun ifStmt(condition: Expr, block: ProgramBuilder.() -> Unit): IfBuilder
 }
 
-fun ProgramBuilder.local(name: String): Local = Local(name)
-fun ProgramBuilder.param(index: Int): Parameter = Parameter(index)
-fun ProgramBuilder.thisRef(): Expr = ThisRef
-fun ProgramBuilder.const(value: Double): Constant = Constant(value)
+fun ProgramBuilder.local(name: String) = Local(name)
+fun ProgramBuilder.param(index: Int) = Parameter(index)
+fun ProgramBuilder.thisRef() = ThisRef
+fun ProgramBuilder.const(value: Int) = ConstantNumber(value.toDouble())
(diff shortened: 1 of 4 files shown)


jacodb-ets/src/main/kotlin/org/jacodb/ets/dsl/ProgramBuilder.kt
```diff
@@ -18,35 +18,60 @@
...
+fun ProgramBuilder.const(value: Int) = ConstantNumber(value.toDouble())
+fun ProgramBuilder.const(value: Double) = ConstantNumber(value)
+fun ProgramBuilder.const(value: Boolean) = ConstantBoolean(value)
+fun ProgramBuilder.const(value: String) = ConstantString(value)
+
+fun ProgramBuilder.and(left: Expr, right: Expr) = BinaryExpr(BinaryOperator.AND, left, right)
+fun ProgramBuilder.or(left: Expr, right: Expr) = BinaryExpr(BinaryOperator.OR, left, right)
+fun ProgramBuilder.eq(left: Expr, right: Expr) = BinaryExpr(BinaryOperator.EQ, left, right)
+fun ProgramBuilder.neq(left: Expr, right: Expr) = BinaryExpr(BinaryOperator.NEQ, left, right)
+fun ProgramBuilder.eqq(left: Expr, right: Expr) = BinaryExpr(BinaryOperator.EQQ, left, right)
```
CaelmBleidd, line 44: Odd naming

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/BlockCfgBuilder.kt
```diff
@@ -109,6 +112,164 @@ class EtsBlockCfgBuilder(
...
+                return entity
+            }
+            return ensureLocal(entity)
+        }
+
+        fun Expr.toEtsEntity(): EtsEntity = when (this) {
+            is Local -> {
+                EtsLocal(
+                    name = name,
+                    type = EtsUnknownType, // TODO
```
CaelmBleidd, line 126: I think you can copy the type?

CaelmBleidd (approved):

jacodb-ets/src/main/kotlin/org/jacodb/ets/dsl/ProgramBuilder.kt
Lipen: I though I have seen such naming somewhere else, but now I can't find any mentions of `eqq` in the whole Internet. 

Do you think `strictEq` would be better?

jacodb-ets/src/main/kotlin/org/jacodb/ets/utils/BlockCfgBuilder.kt
Lipen: DSL's `Local` does not have `type`.

Lipen: I do not think we should make DSL much more complex than it is right now. If you need "typed" local, you can now use `CustomValue { EtsLocal("name", type) }`.

PR #349, opened 2025-08-19
Add literal types

Lipen:
This PR adds proper support for different TS literal types: string/number/boolean. Each literal class is now represented as a separate `data class` under the `sealed interface EtsLiteralType`.

### DIFF
--- a/jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Type.kt
+++ b/jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Type.kt
@@ -41,7 +41,11 @@ interface EtsType : TypeName, CommonType {
         fun visit(type: EtsUndefinedType): R
         fun visit(type: EtsVoidType): R
         fun visit(type: EtsNeverType): R
-        fun visit(type: EtsLiteralType): R
+
+        // Literal
+        fun visit(type: EtsStringLiteralType): R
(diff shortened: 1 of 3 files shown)


jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Type.kt
```diff
@@ -277,11 +284,39 @@ object EtsNeverType : EtsPrimitiveType {
...
-data class EtsLiteralType(
-    val literalTypeName: String,
-) : EtsPrimitiveType {
+sealed interface EtsLiteralType : EtsPrimitiveType
+
+data class EtsStringLiteralType(
+    val value: String,
+) : EtsLiteralType {
+    override val typeName: String
+        get() = "\"$value\""
```
CaelmBleidd, line 293: Are this `"` necessary? Probably they will be a noise we will filter in usages

CaelmBleidd (approved):

jacodb-ets/src/main/kotlin/org/jacodb/ets/model/Type.kt
Lipen: https://github.com/UnitTestBot/jacodb/blob/neo/jacodb-api-jvm/src/main/kotlin/org/jacodb/api/jvm/cfg/JcInst.kt#L1184

